PR #1, opened 2021-04-22
Add multi-order coverage (MOC) support to pgsphere

msdemlei:
This PR, developed by Credativ in https://github.com/credativ/pgsphere/, adds support for IVOA Multi-Order Coverage maps (https://ivoa.net/documents/MOC/). They can be serialised and deserialised (to ASCII MOCs), they can be produced from polygons and circles, the can be intersected and joined. A simple indexing scheme is also provided.


obartunov:
Thanks for PR, we will test it.

PR #6, opened 2021-12-09
Add MOC vs. other geometries casts

msdemlei:
This PR will enable search conditions such as scircle <@ moc.  It is implemented as syntactic sugar, turning this, in effect, into smoc(max_order(moc), scircle) <@ moc.

While this has several drawbacks against a strict implementation of contains and overlap between spherical geometries and mocs – most notably, it will be a lot slower when comparing a constant geometry against moc-valued columns, because the conversion will be re-done in every row –, implementing these exact operations is at least hard.

For casual users, I argue, this will provide non-surprising behaviour.

This PR also bumps the version to 1.2.1.  Piggyback on it rides a brief hacking guide I came up with while writing this.


msdemlei:
Can I bump this?  The CI failures, I think, are unrelated to the PR, as pgsphere probably doesn't work with 9.x anyway (or does it?  And if so, is there a strong case to not drop 9.x support?).

I give you this isn't perfect and rather certainly won't be terribly fast, but then it *will* help for tables in the ~1e4 rows range, such as rr.stc_spatial in future RegTAP (the VO registry discovery protocol, https://github.com/ivoa-std/RegTAP).

msdemlei:
Since it'd be great to have these operators for RegTAP 1.2, https://ivoa.net/documents/RegTAP/20220519/, can I nag again about this?  I realise we'll not win beauty contests with this, but I'd say it's reasonably harmless and cheap while giving a rather real benefit...

gmantele (approved):
It seems all right to me.
It passes all CI for Postgres 10 and more.

PR #14, opened 2023-06-20
Source files are moved into src subdirectory.

vitcpp:
Source files were moved into the src subdirectory in order to avoid file mess in the project root directory.


vitcpp:
@gmantele, @msdemlei Could you please review the patch. Is it suitable for you?

vitcpp:
It seems some checks are failed. I will recheck it.

Makefile
```diff
@@ -4,11 +4,11 @@ PGSPHERE_VERSION = 1.2.1
...
 MODULE_big = pg_sphere
-OBJS       = sscan.o sparse.o sbuffer.o vector3d.o point.o \
-             euler.o circle.o line.o ellipse.o polygon.o \
-             path.o box.o output.o gq_cache.o gist.o key.o \
-             gnomo.o healpix.o moc.o process_moc.o healpix_bare/healpix_bare.o \
-             epochprop.o
+OBJS       = src/sscan.o src/sparse.o src/sbuffer.o src/vector3d.o src/point.o \
+             src/euler.o src/circle.o src/line.o src/ellipse.o src/polygon.o \
+             src/path.o src/box.o src/output.o src/gq_cache.o src/gist.o src/key.o \
+             src/gnomo.o src/healpix.o src/moc.o src/process_moc.o healpix_bare/healpix_bare.o \
```
esabol, line 10: I would wrap this line, but that's a minor nit.

vitcpp: @esabol thank you for the review. I've added one more commit with wrapping some lines. I've also aligned identation of some lines.

esabol: Well, the indentation of lines 8 and 9 look different from the indentation of lines 10 and 11 now. Are you using a different number of tabs on these lines to indent?

vitcpp: @esabol Finally, I've decided to wrap the long line and nothing more. All my previous explanations seems to be wrong because postgresql coding standards are related to source files. Sorry. I propose to reformat Makefile in a separate PR later because there are different formatting styles used inside - the mix of spaces and tabs.

esabol: > What do you think?

My comment was just that lines 8-11 should be consistent in how they are indented.

I _think_ tabs are expected to be 8 spaces wide in Makefiles, but I'm not 100% sure.

Continuation lines like this should be indented with at least one tab.

vitcpp: @esabol I think I've done what you proposed. Now indentation is consistent for these lines. I apologize I have made some noise here. We can go further If everything is ok now. Thank you!

esabol: It looks perfect to me, @vitcpp ! Thank you!

esabol (approved):

msdemlei:
On Tue, Jun 20, 2023 at 05:27:22AM -0700, Vitaly wrote:
> @gmantele, @msdemlei Could you please review the patch. Is it
> suitable for you?

Builds for me, too (gcc 10.2.1 and 12.2.0), the result passes the
tests, and the source tree looks quite a bit neater.  Thanks!

PR #23, opened 2023-06-27
[ISSUE-10] Implemented propagation of pg sphere version.

vitcpp:
Implemented propagation of pg sphere version from Makefile into the code. Fixed pg_sphere_version function.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,9 @@ TESTS       = init_test tables points euler circle line ellipse poly path box in
 			  contains_ops contains_ops_compat bounding_box_gist gnomo healpix \
 			  moc mocautocast epochprop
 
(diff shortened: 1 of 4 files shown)


vitcpp:
It seems there are build problems on 11+ versions. The error is occured when compiling with clang. It seems PG_CFLAGS is not utilised with clang. Not sure, how clang compilation is performed. I see no clang rules. It is interesting that there is no clang compilation with 10 version. Investigating it.

vitcpp:
The problem with clang was because of JIT compilation (--with-llwm option). Variables PG_CXXFLAGS, PG_CFLAGS are not propagated into clang compilation. Using PG_CPPFLAGS fixes the problem.

esabol (approved):

vitcpp:
@esabol thank you for the review. I'm going to merge this PR because there are no objections.

PR #25, opened 2023-07-03
[ISSUE-7] Added PARALLEL SAFE mark to pgSphere functions

vitcpp:
The patch is based on Feodor Sigaev's branch (parallel_safe). Some adjustments were made to fix compilation. Checked the number of changes in function declarations and the number of function alterations in the upgrade script pg_sphere--1.2.1--1.2.2.sql.in (the numbers should match).


Makefile
```diff
@@ -209,6 +210,12 @@ ifeq ($(has_parallel), n)
 	sed -i -e '/PARALLEL/d' $@ # version $(pg_version) does not have support for PARALLEL
 endif
 
+pg_sphere--1.2.1--1.2.2.sql: upgrade_scripts/pg_sphere--1.2.1--1.2.2.sql.in
+	cat $^ > $@
+ifeq ($(has_parallel), n)
+	sed -i -e '/PARALLEL/d' $@ # version $(pg_version) does not have support for PARALLEL
```
esabol, line 216: Just checking, but shouldn't this line (and lines 120 and 204 and 210) be `sed -i -e '/ *PARALLEL SAFE/d' $@`?

vitcpp: These lines work only for pg versions less than 9.6 (PARALLEL SAFE is not implemented for earlier versions). Now I removed 9.X versions from tests because these versions are not supported by the community. Furthermore, I'm not sure that sed is executed correctly. It removes the whole line, not only parallel safe marks (MMUTABLE STRICT seems to be removed as well). When I did the patch I just 'copy-pasted' the old solution. I propose to keep my patch unchanged but to rework these lines in another Issue if we still need support of 9.X versions.

esabol: I think the sed commands should be `sed -i -e 's/  *PARALLEL SAFE//' $@`, but I agree there's no point in continuing to support Pg 9.x, so might as well just remove all of the sed commands.

esabol:
By the way, if you put "[ISSUE #7]" in the title of the PR instead of "[ISSUE-7]", GiHub will automatically link this PR to that issue.

Makefile
vitcpp: @esabol I agree, I think to remove obsolete lines of code. I will do it then.

esabol (approved):

vitcpp:
@esabol Thank you for the review!

Issue #19, opened 2023-06-23
Make healpix+smoc compilation optional


vitcpp:

esabol:
Yeah, I could really use this, I think. Our main database server is still RHEL 6.x, and I don't think there's a libhealpix for it. I'm not seeing a package for it at least.

I'm getting compilation errors when trying to compile the current master branch of this repo:
```
g++ -Wall -Wpointer-arith -Wendif-labels -Wmissing-format-attribute -Wformat-security -fno-strict-aliasing -fwrapv -O2 -fPIC -DPGSPHERE_VERSION=1.2.1 -I. -I./ -I/local/postgres/install/15.3/include/server -I/lo
cal/postgres/install/15.3/include/internal  -D_GNU_SOURCE    -c -o src/process_moc.o src/process_moc.cpp
src/process_moc.cpp:17:26: error: healpix_base.h: No such file or directory
src/process_moc.cpp: In function 'size_t get_moc_debug(const char**, void (*)(const char*, int))':
src/process_moc.cpp:69: error: expected type-specifier before 'PlanckError'
src/process_moc.cpp:69: error: expected ')' before '&' token
src/process_moc.cpp:69: error: expected '{' before '&' token
src/process_moc.cpp:69: error: 'e' was not declared in this scope
src/process_moc.cpp:69: error: expected ';' before ')' token
src/process_moc.cpp:69: error: expected primary-expression before 'catch'
src/process_moc.cpp:69: error: expected ';' before 'catch'
src/process_moc.cpp: In function 'void moc_debug_log(void (*)(const char*, int), const char*, ...)':
src/process_moc.cpp:84: error: expected type-specifier before 'PlanckError'
src/process_moc.cpp:84: error: expected ')' before '&' token
src/process_moc.cpp:84: error: expected '{' before '&' token
src/process_moc.cpp:84: error: 'e' was not declared in this scope
src/process_moc.cpp:84: error: expected ';' before ')' token
src/process_moc.cpp:84: error: expected primary-expression before 'catch'
```
etc.

Any short-term workarounds that I can employ?

esabol:
> Any short-term workarounds that I can employ?

I hacked away at the Makefile using https://github.com/postgrespro/pgsphere/pull/13 as a guideline and managed to get it to compile, install, and crushtest OK. Whew!

vitcpp:
[...]

PR #26, opened 2023-07-06
[ISSUE #19] Make HEALPix/MOC support optional

esabol:
This PR address issue #19 and makes support for HEALPix/MOC functionality optional. It is enabled by default, but it can be disabled by including `USE_HEALPIX=0` as an argument to the `make` commands. The README.pg_sphere file is updated to reflect this change (and other recent changes to the build process).

Also, when upgrading my installation from 1.2.1 to 1.2.2, I was getting the error:
```
# ALTER EXTENSION pg_sphere UPDATE TO '1.2.2';
ERROR:  function g_spoint3_fetch(internal, internal, internal) does not exist
```
I've included a fix for that in this PR (the function is just `g_spoint3_fetch(internal)`). Let me know if you prefer that I submit that as a separate PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -235,7 +235,7 @@ pg_sphere--1.1.5beta4gavo--1.2.0.sql: pgs_moc_ops.sql.in
 	cat upgrade_scripts/$@.in $^ > $@
 
 pg_sphere--1.2.0--1.2.1.sql: pgs_moc_geo_casts.sql.in pgs_epochprop.sql.in
-	cat $^ > $@
+	cat upgrade_scripts/$@.in $^ > $@
 
 pg_sphere--1.2.1--1.2.2.sql: upgrade_scripts/pg_sphere--1.2.1--1.2.2-healpix.sql.in
 	cat upgrade_scripts/$@.in $^ > $@
--- a/Makefile
+++ b/Makefile
@@ -134,6 +134,10 @@ else
 endif
 
 test: pg_sphere.test.sql sql/init_test.sql
+	cp expected/init_test.out.in expected/init_test.out
+ifneq ($(USE_HEALPIX),0)
+	cat expected/init_test_healpix.out.in >> expected/init_test.out
+endif
 	$(pg_regress_installcheck) $(PGS_TMP_DIR) $(REGRESS_OPTS) $(TESTS)
 
 pg_sphere.test.sql: $(RELEASE_SQL) $(shlib)
--- a/Makefile
+++ b/Makefile
@@ -30,24 +30,21 @@ DATA_built  = $(RELEASE_SQL) \
 
 DOCS        = README.pg_sphere COPYRIGHT.pg_sphere
 REGRESS     = init tables points euler circle line ellipse poly path box index \
-              contains_ops contains_ops_compat bounding_box_gist gnomo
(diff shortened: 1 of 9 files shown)


esabol:
`make test` fails, but I don't think it worked before this PR either? Does it work for you? Does anyone care? `make crushtest` works fine...

vitcpp:
@esabol Nice catch! You've found g_spoint3_fetch issue. I propose to do it in a separate PR that will be merged prior to your main PR. It is because I would like to propose to start assigning version tags. Thank you!

vitcpp:
@esabol Concerning make test. It fails without your PR. I will create an issue for it.

Makefile
```diff
@@ -24,14 +30,24 @@ DATA_built  = $(RELEASE_SQL) \
...
 REGRESS     = init tables points euler circle line ellipse poly path box index \
-			  contains_ops contains_ops_compat bounding_box_gist gnomo healpix \
-			  moc mocautocast epochprop
+              contains_ops contains_ops_compat bounding_box_gist gnomo
+
+ifneq ($(USE_HEALPIX),0)
+REGRESS    += healpix moc mocautocast
+endif
+
+REGRESS    += epochprop
```
vitcpp, line 39: I propose not to separate epochprop into a new line. It is already included unconditionally.

esabol: I didn't want to change the order of the tests, as I thought that might affect something. I think I saw a comment somewhere saying that "the order matters" and not to change it, but maybe that doesn't apply to `REGRESS`. If it's ok to change the order of the tests, it would certainly simplify things if I could add it unconditionally above this line and I will update the PR.

vitcpp: I may guess that epochprop independent on other tests. I tried to move epochprop back and forth and I haven't seen any problems with it. I got your point. It is up to you to keep PR unchanged. Furthermore, we are planning to cleanup and improve Makefile in the future. I do not see any other problems with your PR at the moment.

esabol:
With the help of Travis CI, I got `make test` working again, both with HEALPix and without. Some of the line numbers were just off by one. I've included these changes in this PR since the expected output for `make test` is different with `USE_HEALPIX=0` and without, so I think it's related. This work addresses issue #28 as a result. I think we should continue to keep `init_test.sql`. I think it's meant to be used to test the build _before_ you install it. So perhaps the README should recommend that the procedure should be `make; make test; make install; make installcheck; make crushtest`?

I've also added `make test` and `make crushtest` to the Travis CI and added a separate `USE_HEALPIX=0` build and it's testing. All this adds approximately 1 minute to each CI job or about 6 minutes for all builds in PR. I mention this since I know Travis CI costs money. These changes address issue #27.

That just leaves the minimum PostgreSQL version mentioned in the `README.pg_sphere` file. I've changed it to version 9.6 in the most recent commits. Other than that, this PR is ready to merge if you like what you see. Thanks!

vitcpp:
@esabol You did great work! Give me please some time to review the changes. Thank you!

vitcpp:
@esabol I tested your changes. It works for me with and without USE_HEALPIX. Thank you!

vitcpp (approved):

esabol:
Thank you, @vitcpp !

Issue #17, opened 2023-06-23
strans is broken


vitcpp:
There is an issue with strans that was reported in another repo. Check and fix it in our repo.

https://github.com/akorotkov/pgsphere/issues/11

esabol:
Could be related to this precision issue?

https://github.com/akorotkov/pgsphere/issues/13

vitcpp:
I think double precision should be ok. I haven't looked at this problem yet. I think it may be related to the custom epsilon that is choosen as 1E-9 in pgsphere (pg_sphere.h:48). May be it is the source of the problem. May be we should make EPSILON smaller.

stepan-neretin7:
Issue can be closed, probably, @vitcpp

vitcpp:
@stepan-neretin7 Thank you for your contribution. Closing the issue.

PR #32, opened 2023-07-21
[ISSUE #17] Corrected the angle functions to fix the incorrected results strans function according to the documentation.

stepan-neretin7:
The problem was that our corners are float8, and for some reason pointers were returned.


esabol (approved):

vitcpp (approved):

vitcpp:
I propose to change version to 1.2.3 (increment patch number) before merge.

@stepan-neretin7 Could you please change the version number to 1.2.3 in the following places:
- Makefile: change PGSPHERE_VERSION to 1.2.3
- pg_sphere.control: change default_version to 1.2.3

sql/euler.sql
```diff
@@ -1,9 +1,55 @@
...
+
+SELECT phi( - strans ( '20d, 50d, 80d, XYZ' ) );
+
+SELECT strans( '90d, 60d, 30d' );
+
+SELECT theta( strans( '90d, 60d, 30d' ) );
+
+SELECT psi( strans( '90d, 60d, 30d' ) );
+
+SELECT phi( strans( '90d, 60d, 30d' ) );
```
esabol, line 55: You're missing a linefeed at the end of this file. This is a POSIX requirement.

esabol (changes requested):

sql/euler.sql
stepan-neretin7: fixed it

vitcpp:
@stepan-neretin7 Thank you for the fix! It works on my side. I think the patch is ready for merge.

The decision to increment version number was made because the fix is important. I propose to define version 1.2.3 as stable and assign the tag '1.2.3'. Stub upgrade script was added to make upgrade process to complete without errors.

esabol (approved):

esabol:
Looks good to me! Thank you!

Issue #16, opened 2023-06-23
Implement SQL functions to access points in spoly, spath by index and as an array


vitcpp:
It seems existing SQL API misses such functions. Check and implement it if absent.

vitcpp:
Closed as compeleted

PR #34, opened 2023-07-27
[ISSUE #16] Implemented a function and tests to extract vertices from spoly by index 

stepan-neretin7:


src/polygon.c
```diff
@@ -554,6 +557,58 @@ spoly_segment(SLine *sl, const SPOLY *poly, int32 i)
...
+            // Clean up and return NULL
+            for (size_t j = 0; j < i; j++)
+                pfree(DatumGetPointer(datum_arr[j]));
+            pfree(datum_arr);
+            PG_RETURN_NULL();
+        }
+        datum_arr[i] = PointerGetDatum(&p[i]);
+    }
+
+    res = construct_array(datum_arr, poly->npts, TypenameGetTypid("spoint"), sizeof(SPoint), false, 'd');
```
vitcpp, line 607: TypenameGetTypid should not be called every time when spherepoly_get_array is called because it is time consuming operation. The idea is to cache spoint type in the process memory (in a static variable) id and reuse it.

Issue #35, opened 2023-07-29
spherepoly_check (sline_sline_pos) is not working properly


ggnmstr:
Current state of spherepoly_check allows user to create polygons like spoly'{(0d,1d),(0d,2d),(0d,3d)}', which is incorrect because all points lie on the same meridian.
The root of this problem may be in sline_sline_pos fucntion, because it determines relations between  sline'(0d,1d),(0d,2d)'  and sline'(0d,3d),(0d,1d)' as "connects", which is also incorrect. 

![Screenshot from 2023-07-29 15-23-52](https://github.com/postgrespro/pgsphere/assets/61929074/defeeada-6d70-489f-91a7-e180a0c2e1de)
2 last log messages: pos=5 means lines are connected, but they actually overlap.

Solution I found is to change the order of block of code in sline_sline_pos responsible for detecting overlap and connection. 
I created pull request with that fix.

Issue #22, opened 2023-06-23
Add spoint_deg() and spoly_deg() functions?


esabol:
All of the database tables we use have the spherical positions stored as degrees, not radians. Adding spoint_deg() and spoly_deg() would be useful for this use-case. There's a 5-year-old PR that implements this here:

https://github.com/akorotkov/pgsphere/pull/6

It just needed a little clean up, but nobody followed through. Would you consider incorporating this functionality here?

vitcpp:
@esabol Sure, I will check and incorporate it. I know about akorotkov repo but I'm not sure that it still alive. Anyway, if someone wants to create a ready-to-merge PR, it will make the process faster (assign yourself on the Issue first to avoid conflicts). There are a lot of other pending PR. I plan to review and incorporate some other pending PR in our branch.

stepan-neretin7:
Maybe we can use set_sphere_output to implement this functions? @esabol

vitcpp:
@stepan-neretin7 It is important for some queries to have functions that return angles in different units independently on a global context. For example, you want to output a table with two columns in radians and in degrees and do some calculations in both degrees and in radians in the same query. set_sphere_output is used to configure how the objects are displayed in user console.

vitcpp:
@stepan-neretin7 set_sphere_output is used to configure how objects are displayed in user console. spoint_deg and spoly_deg construct new objects. Users may use radians and degrees to construct different objects in the same query. Of course, user may use textual representation of coordinates in degrees by adding 'd' suffix but it is not comfortable because user have to format a string.

esabol:
The typical use-case is that you have a database table with two columns containing longitudes and latitudes in degrees. I think most everyone stores longitudes and latitudes in degrees. Fortunately, PostgreSQL has a radians() function to convert our degrees to the radians that pgSphere requires. So a typical pgSphere use-case is a query like this:
```
select * from sometable where spoint(radians(ra),radians(dec)) @ scircle(spoint(radians(10),radians(10)), radians(1))
```
Well, adding all those `radians()` functions to your SQL is cumbersome. Also, it would be faster and more efficient if pgSphere's most commonly used functions had alternate versions which take degrees and convert to radians internally (at C code level) instead of using PostgreSQL's `radians()` function. In such a scenario, the SQL then becomes:
```
select * from sometable where spoint_deg(ra,dec) @ scircle_deg(spoint_deg(10,10),1))
```
which is both more succinct and probably faster.

[...]

PR #36, opened 2023-07-29
Fix sline_sline_pos

ggnmstr:
The problem fixed by this pull request is described in Issue #35

### DIFF
--- a/expected/poly.out
+++ b/expected/poly.out
@@ -350,6 +350,15 @@ SELECT spoly '{(0d,0d),(10d,10d),(0d,10d),(10d,0d)}';
 ERROR:  spherepoly_from_array: a line segment overlaps or polygon too large
 LINE 1: SELECT spoly '{(0d,0d),(10d,10d),(0d,10d),(10d,0d)}';
                      ^
+--- degenerate polygons -----
+SELECT spoly '{(0d,1d),(0d,2d),(0d,3d)}';
+ERROR:  spherepoly_from_array: a line segment overlaps or polygon too large
+LINE 1: SELECT spoly '{(0d,1d),(0d,2d),(0d,3d)}';
+                     ^
+SELECT spoly '{(1d,0d),(2d,0d),(3d,0d)}';
+ERROR:  spherepoly_from_array: a line segment overlaps or polygon too large
+LINE 1: SELECT spoly '{(1d,0d),(2d,0d),(3d,0d)}';
+                     ^
 --- functions
 SELECT npoints( spoly '{(10d,0d),(10d,1d),(15d,0d)}');
  npoints 
--- a/sql/poly.sql
+++ b/sql/poly.sql
@@ -96,6 +96,12 @@ SELECT spoly_deg(NULL::float8[]);
 
 SELECT spoly '{(0d,0d),(10d,10d),(0d,10d),(10d,0d)}';
 
+--- degenerate polygons -----
+
+SELECT spoly '{(0d,1d),(0d,2d),(0d,3d)}';
+
+SELECT spoly '{(1d,0d),(2d,0d),(3d,0d)}';
+
 --- functions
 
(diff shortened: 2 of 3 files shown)


esabol:
Always connect your pull requests to their relevant issue numbers, please. You can do this by mentioning the issue number in the title or description of the PR.

Issue #35

ggnmstr:
Sure, I added a pair of tests to sql/poly.sql
The first one definitely passes with the PR and fails without it (both polygons shouldn't be created).

esabol:
Looks good to me! Thanks, @ggnmstr!

What do you think, @vitcpp?

esabol (approved):

vitcpp:
Colleagues, I have some doubts about this change. I think, degenerate polygons can be accepted by most of the algorithms. This PR disables some types of polygons that may affect on user experience - if the old version constructed such degenerate polygons, the new version reports an error. Some user datasets may contain degenerate polygons. Explicit error reporting may stop data pipelines to process some datasets. Give me please some more time to think about it.

esabol:
> This PR disables some types of polygons that may affect on user experience - if the old version constructed such degenerate polygons, the new version reports an error. Some user datasets may contain degenerate polygons. Explicit error reporting may stop data pipelines to process some datasets. Give me please some more time to think about it.

If there's a use-case for supporting degenerate polygons, I'd be open to it, but it seems to me that giving correct answers is what should matter. Degenerate polygons are not valid polygons, according to the definition given by the [OGC Simple Features Specification](http://www.opengeospatial.org/standards/sfa). See https://cs.stackexchange.com/a/112703.

ggnmstr:
I found out that PGSphere is kind of inconsistent in terms of degenerate polygons. It doesn't allow to create all types of degenerate polygons:
before-patch
Meanwhile PR I suggest removes some possibilities to create degenerate  polygons (connected by 0d coordinate), but not all:
after-patch
So it may be some kind of problem with these 0d coordinates, maybe math, not sure now. 
But first, of course, we need to decide whether degenerate polygons should be allowed or not.

esabol:
Interesting. Thank you for testing that and reporting your findings, @ggnmstr.

vitcpp:
It seems there is a more complex issue with degenerate polygons. Thank you @ggnmstr  for your findings. There is the working idea to separate two concepts - storage and algorithms. spoly is a storage for polygon data. The storage may or may not validate the input data because in most cases the storage just used to manipulate data but not to validate it. My position is to allow to keep both valid and invalid data in storage. Data import algorithms just load data into storage objects as is. Some functions to validate and normalize data should be implemented as well. The algorithms assume that the input storage objects contain valid data. Such check is not trivial for polygons. It is the developer responsibility to pass valid data to algorithms. Otherwise, the behaviour is undefined. For most of the algorithms degenerate polygons are ok but it is important to avoid edge crosses that may lead to the wrong results. Overlapping edges may be used to construct polygons with holes (if needed).

Now we have some inconsistent behaviour with spoly. Once we can't create degenerate spoly with 4 points and it is the adopted behaviour I think I'm ok to disable creation of all types of degenerate polygons. Later, if we decide to allow degenerate polygons we can change this behaviour.

Lets wait for PR #22 and then merge it.

vitcpp:
@ggnmstr Could you please rebase the branch to the newer version and start the comment from the capital letter? We will merge it if tests are passed.

vitcpp (changes requested):
@ggnmstr Could you, please, start the commit description from the capital letter?

ggnmstr:
@vitcpp Done

vitcpp (approved):
Once the tests are passed, I approve it. One trivial request is to start the PR title with a capical letter. @ggnmstr Thank you!

ggnmstr:
Done. 
@esabol, @vitcpp, thank you for review.

PR #39, opened 2023-07-31
Fix compatibility with PG16

df7cb:
* Replace Abs with fabs
* Remove DatumGetPointer where PG16 got stricter typechecking
* Annotate expected/moc100 output files with PG major and architecture bits they are meant for
* Add expected output files for moc100 on PG16


esabol:
Would it be premature to add PostgreSQL 16 to the Travis CI? Does `apt.postgresql.org.sh` support PostgreSQL 16 yet? If 16 does work, then I suggest adding it in this PR.

Just curious, where did `Abs()` from in the first place?

vitcpp:
I'm not sure how the reference output file moc100_x.out is chosen for comparison. Is it performed automatically by test scripts?

df7cb:
> I'm not sure how the reference output file moc100_x.out is chosen for comparison. Is it performed automatically by test scripts?

https://www.postgresql.org/docs/current/regress-variant.html

> Would it be premature to add PostgreSQL 16 to the Travis CI? Does `apt.postgresql.org.sh` support PostgreSQL 16 yet? If 16 does work, then I suggest adding it to `.travis.yml` in this PR.

Of course PG16 is already supported. I'll add that shortly.

> Just curious, where did `Abs()` come from in the first place? Is that an older PostgreSQL thing?

Abs was removed in PG16.

vitcpp:
@df7cb Thank you for the explanation. I'm ok with the patch but we have to fix tests for pg16. I'm not sure how to fix it yet. May be to change snapshot to the latest one?

vitcpp:
I propose to enable and fix PG16 tests in another PR and then apply this patch. I can do it but I'm on vacation this week with limited access to the internet.

esabol:
@vitcpp wrote:

> I propose to enable and fix PG16 tests in another PR and then apply this patch. I can do it but I'm on vacation this week with limited access to the internet.

This sounds to me like a very good plan.

Thank you, @df7cb, for detailing the issues with the test suite and such.

vitcpp:
@df7cb We fixed some warning messages that resulted in test failures. Could you please be so kind to re-sync your branch? Thank you in advance!

df7cb:
Branch rebased.

esabol (approved):
Looks good to me! Let's merge. Thank you , @df7cb !

PR #44, opened 2023-08-04
Created some tests for contains and overlap operations to check DE-9IM compliance

3Davydov:


esabol:
Thank you for this, @3Davydov ! I love more tests.

There's already a "contains_ops" and a "contains_ops_compat" files for testing. I wonder if it would be appropriate to put your "contains" tests in one of those files?

3Davydov:
Hello, @esabol !
I think it's still worth making a separate file for this. The tests in the "contains_ops" and "contains_ops_compat" files are relevant only for the current logic.
If it is decided to switch to the DE-9IM model, then it will be necessary first of all to focus on the tests from "contains", and it may be more convenient to work with a separate file.

esabol (approved):

sql/contains.sql
```diff
@@ -0,0 +1,123 @@
...
+select 'sbox ~ spoint', 'f' as expected, sbox'( (0d,0d), (20d,10d) )' ~ spoint'(0d, 0d)' as actual;
+
+-- the point lies in the interior of box
+select 'sbox ~ spoint', 't' as expected, sbox'( (0d,0d), (20d,10d) )' ~ spoint'(5d, 5d)' as actual;
+
+-- the point and box that degenerated into the point coincide
+select 'sbox ~ spoint', 't' as expected, sbox'( (0d,0d), (0d,0d) )' ~ spoint'(0d, 0d)' as actual;
+
+-- the point and box that degenerated into the point do not coincide
+select 'sbox ~ spoint', 'f' as expected, sbox'( (0d,0d), (0d,0d) )' ~ spoint'(0d, 1d)' as actual;
```
vitcpp, line 123: Add newline symbol

sql/overlaps.sql
```diff
@@ -0,0 +1,484 @@
...
+select 'sbox && sellipse', 't' as expected, sbox'((-10d , -10d), (10d , 10d))' && sellipse'<{ 10d , 5d }, (10d, 10d) , 90d>' as actual;
+
+-- the box degenerated into the point lies in the boundary of ellipse
+select 'sbox && sellipse', 'f' as expected, sbox'((30d , 0d), (30d , 0d))' && sellipse'<{ 30d , 20d }, (0d, 0d) , 0d>' as actual;
+
+-- the box degenerated into the point lies in the interior of ellipse
+select 'sbox && sellipse', 'f' as expected, sbox'((0d , 0d), (10d , 0d))' && sellipse'<{ 30d , 20d }, (0d, 0d) , 0d>' as actual;
+
+-- the box and ellipse are degenerated into the point and coincide
+select 'sbox && sellipse', 'f' as expected, sbox'((0d , 0d), (0d , 0d))' && sellipse'<{ 0d , 0d }, (0d, 0d) , 0d>' as actual;
```
vitcpp, line 484: Add newline symbol

vitcpp (changes requested):
@esabol I propose to accept this PR with some trivial changes. Tests do not affect functionality but can be used to learn semantics of operations. I have some ideas how to improve and reorganise it. @3Davydov did a nice analysis of existing algorithms and found that they are not fully compliant with the semantics of operations on figures as described in the DE-9IM model (and OSG Simple Feature Specification). I understand that we can not just change the behaviour of operations to make it compliant but we can update the docs with some notes about differences with reference models.

sql/contains.sql
esabol: Yes, POSIX requires newline at the end of all text files.

3Davydov: Done

sql/overlaps.sql
3Davydov: Done

esabol (approved):

vitcpp (approved):

PR #45, opened 2023-08-04
Updated the current version.

vitcpp:
It seems we forgot to update the current version when merging issue #22. This patch updates the current version to 1.3.0.


esabol (approved):
Yes, of course. Thank you!

esabol (approved):

Issue #42, opened 2023-08-02
Build warning in process_moc.cpp


esabol:
This build warning in process_moc.cpp should be easy to fix, I hope:
```
src/process_moc.cpp: In function ‘void create_moc_release_context(void*, Smoc*, pgs_error_handler)’:
src/process_moc.cpp:624:17: warning: declaration of ‘n’ shadows a previous local [-Wshadow=compatible-local]
  624 |    rnode_iter n(moc_data, m.layout[k + 1].level_end);
      |                 ^~~~~~~~
src/process_moc.cpp:580:14: note: shadowed declaration is here
  580 |   rnode_iter n(moc_data, m.layout[1].level_end);
      |              ^
```

stepan-neretin7:
Can we closed it?

esabol:
PR https://github.com/postgrespro/pgsphere/pull/47 has been merged, addressing this issue. Closing.

PR #47, opened 2023-08-04
[ISSUE #42] Fix compiler warning

stepan-neretin7:


esabol (approved):
Looks good to me! Thank you, @stepan-neretin7 !

vitcpp (approved):
In general, I'm ok with the fix. One trivial concern is the chosen variable name, but I do not see the sense in asking authors about it. It may be easily changed in the future.

PR #49, opened 2023-08-09
Fix pgSphere version and the project url in the doc

vitcpp:


esabol (approved):
It would be nice if the version number could be set automatically based on the value in the Makefile, but this is fine for now.

vitcpp:
@esabol  Yes, sure. Thank you. This is a fast fix to make the doc consistent. Propagation of the version number will be done in a separate task to review and improve the document generation.

PR #51, opened 2023-08-13
Fix warnings in output.c

vitcpp:
The variable sphere_output_precision type is changed to int because sprintf requires this argument to be of type int when used as the precision specificator (argument for '*' symbol in the sprintf template).

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -21,7 +21,11 @@
  * Default is radians.
  */
 static unsigned char sphere_output = OUTPUT_RAD;
-static short int sphere_output_precision = DBL_DIG;
+
+/*
+ * Defines the precision of floating point values in output.
+ */
+static int sphere_output_precision = DBL_DIG;
 
 PG_FUNCTION_INFO_V1(set_sphere_output);
 PG_FUNCTION_INFO_V1(spherepoint_out);


src/output.c
```diff
@@ -21,7 +21,11 @@
  * Default is radians.
  */
 static unsigned char sphere_output = OUTPUT_RAD;
-static short int sphere_output_precision = DBL_DIG;
+
+/*
+ * Define the precision of output floating point values.
+ */
+static int sphere_output_precision = DBL_DIG;
```
esabol, line 28: I still don’t understand how changing this from short int to int fixes the compiler error:
```
src/output.c:424:18: error: ‘%.*g’ directive writing between 1 and 310 bytes into a region of size between 76 and 92 [-Werror=format-overflow=]
  424 |       "%2ud %2um %.*gs",
      |                  ^~~~
src/output.c:424:7: note: assuming directive output of 309 bytes
  424 |       "%2ud %2um %.*gs",
      |       ^~~~~~~~~~~~~~~~~
```
but I do 100% agree that the argument to sprintf should be int. I’m going to just chalk this up to this gcc version not being quite right.

esabol (approved):
This looks fine. The 100 in `buf[100]` should be more meticulously determined, but, if you’re happy with this, then let’s just merge and move on to incorporating PG16 support.

src/output.c
vitcpp: @esabol sprintf takes variadic arguments that is implemented using va_list in C language. sprintf decides how to format the memory chunk with arguments based on the template. I guess, it is like to read raw bytes from a stream and try to interpret them. There is an undefined behaviour if sprintf expects int type (4 bytes), but short int (2 bytes) is passed instead. It seems 2 extra bytes belong to the next argument, depending on the implementation of va_list. It is a real bug that can lead to some unexpected program behaviour and it should be fixed.

The problem with some magic numbers in temporary buffers is not so important. The case with buffer overflow is unlikely in this case. If we want to redesign this behaviour it should be rewritten completely. Some checks for buffer overflows should be implemented.

vitcpp:
@esabol It is interesting, that I tried this: 

sprintf(&buf[0], "%.*gd, (int)sphere_output_precision, ... 

And I found that it doesn't fix the warning on my side. I use gcc 11.3.0. The warning is enabled by -Wformat-overflow. Some heuristic in gcc is used in this case. It seems it somehow defines that sphere_output_precision may be much greater than DBL_DIG. I can't explain why such change doesn't fix the warning but change the type to int fixes it.

vitcpp:
The warnings are fixed in focal dist. The rest will be fixed in PR #39. Lets incorporate this patch first, #39 will be the next.

esabol (approved):

PR #56, opened 2023-08-25
Fix sphereline_circle_pos (uninitialized memory use)

vitcpp:


esabol (approved):
Good catch! Feel free to force merging in spite of unrelated PG16 failure.

vitcpp:
@esabol It seems this PR has the same problem with moc100 under PG16. I agree to merge this patch. The change is trivial.

PR #57, opened 2023-08-25
Fix diff in the test moc100 on PG16

vitcpp:


esabol:
I confused as to why this change is working on PG 10-15. I thought only the PG16 build broke and the test output was unaltered on PG 10-15.

vitcpp:
@esabol expected/moc100_4.out is applied only in case of PG16:

https://www.postgresql.org/docs/current/regress-variant.html

Other "golden" files expected/moc100_*.out are applied for other versions of postgres. You can see that the beginning of each file contains some output that is specific for each version. It plays the role of a filter. For unsuitable versions, it will produce differences and the golden file will be skipped.

esabol:
Oh, interesting!

esabol (approved):

PR #62, opened 2023-09-07
Fix overlaps.sql test fail on 32 bit Debian due to gcc bug 323

vitcpp:
On some 32 bit platforms, there is a gcc bug that makes floating point calculations and comparisons unstable (see the link below). The problem originates in FPU 80 bits registers where double values are not truncated to 64 bit values. When gcc compiles some code with enabled optimizations, the intermediate results may be kept in the FPU registers without truncation to 64 bit values. Extra bits may produce unstable results when comparing the numbers.

The generic solution is to save the intermediate results in the memory where the values are truncated to 64 bit values. It affects the performance but makes the tests stable on all platforms.

PGSPHERE_FLOAT_STORE macro enables storing of intermediate results for FPxx operations in the memory. It is enabled by default for 32 bit platforms. It can be explicitly enabled or disabled in CFLAGS. To enable it for all code the gcc option -ffloat-store may be used as well.

Link to gcc bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=323


esabol (approved):
Looks good to me!

PR #65, opened 2023-09-19
Remove the obsolete text from the doc

vitcpp:


doc/install.sgm
```diff
@@ -7,8 +7,15 @@
...
-                <ulink url="https://github.com/postgrespro/pgsphere"><citetitle>https://github.com/postgrespro/pgsphere</citetitle></ulink>
+                You can download the latest release sources from the
+                <ulink url="https://github.com/postgrespro/pgsphere/releases">
+                    <application>pgSphere</application> homepage
+                </ulink> (the Releases page). The sources can also be downloaded by
+                cloning the repository with the appropriate release tag (each release is
+                marked with a tag).
+                The master branch is intended for development use
+                and can contain an intermediate work.
+                It is not recommented for use in production.
```
esabol, line 18: Typo. Change "recommented" to "recommended".

esabol (changes requested):

esabol:
While you are at it, could you also remove the `USE_PGXS=1` from the `make` commands on lines 75 and 81 of `doc/install.sgm`? That's completely outdated and unnecessary. 

Setting `PG_CONFIG` still works, but that's mostly unnecessary as well.

df7cb:
I agree that setting `PGXS` on the command line is obsolete since the Makefile already contains it, but setting `PG_CONFIG` is still the primary interface to selecting the target PG version.

`make PG_CONFIG=/usr/lib/postgresql/15/bin/pg_config`

esabol:
Sure, but I feel that's part of the `postgres` user's environment in most cases. At least I always have it set, and I always update that setting whenever I install a new major version release. So I stand by my statement that setting it in the `make` command is "_mostly_ unnecessary", but I have no objection to keeping it in the documentation.

vitcpp:
@esabol I think, in past, there was the scenario when pgsphere was built as a part of postgresql sources tree (contrib/pg_sphere) - lines 88-100 in the top Makefile. Now, pgsphere is not the part of the postgresql tree. I guess, it is why USE_PGXS option was introduced. If we still need to support such scenario we probably should remove USE_PGXS=1 from the makefile and oblige users to set this option in the Makefile as shown in the doc. If we do not need to support such scenario we should cleanup the Makefile, remove obsolete code.

vitcpp:
I think to keep USE_PGXS=1 in the doc because it does nothing. The pg_config is used unconditionally. But, if we decide to support the build in the contribution directory without pg_config, we have to remove the explicit setting of USE_PGXS=1 from the Makefile. It will not affect the developers, who get used to specifying USE_PGXS=1.

df7cb:
Please don't remove USE_PGXS from the Makefile. People would wish to build pgsphere from within a PG source tree can still use `make USE_PGXS=`. But I've never seen anyone do that, so let's not bother the other 99.9% of the users with that flag.

vitcpp:
@df7cb Yes, sure, i'm not going to remove USE_PGXS=1 from the Makefile without discussions and approvals. I just concerned if to put pgsphere into contrib and to run make world, we will have some problems with USE_PGXS=1 flag in the Makefile. It is why I propose to keep USE_PGXS=1 option in the doc. It may be redundand but it may help us to do the changes in the future.

esabol:
> It is why I propose to keep USE_PGXS=1 option in the doc. It may be redundand but it may help us to do the changes in the future.

I disagree. It is useless and confusing because it disagrees with the instructions in README. I have no objection to keeping it in the Makefile, however. I just don't think it should be in the documentation.

I compile pgsphere in my `contrib` directory all the time, and it compiles just fine with `USE_PGXS=1` set in the Makefile.

vitcpp:
@esabol pg_config is used to get the path to the postgresql Makefile.global which implements some common rules. When you set USE_PGXS=1, pg_config at your PATH is used to get the path to Makefile.global. It doesn't matter where you put your pgsphere code.

If you want to compile pgsphere with using of postgresql Makefile.global that is located in your project tree, then you have to set USE_PGXS=0 to disable pg_config because it can return the path to Makefile.global of another installation. In this case, pgsphere Makefile will try to find the postgresql Makefile at path ../..//src/Makefile.global (line 98 of pgsphere Makefile).

pgsphere is not the part of the contrib directory. I'm ok to remove USE_PGXS from the doc. But in this case we have to rewrite the Installation chapter, because there are two build scenarios are described. I propose to keep the only one compilation scenario and to remove the description how to compile in the contrib.

What do you think?

esabol:
Yeah, I understand how the build system works.

> I propose to keep the only one compilation scenario and to remove the description how to compile in the contrib.
>
> What do you think?

Yes, please.

vitcpp:
@esabol I've updated the Installation section. Could you please verify it? Thank you in advance!

doc/functions.sgm
```diff
@@ -25,6 +26,12 @@
 <![CDATA[------]]>
 <![CDATA[ 1]]>
 <![CDATA[(1 row)]]>
+              </programlisting>
+            </example>
+
+            <example>
+              <title>Area of a moc object</title>
```
esabol, line 33: Replace "moc" with "Multi-Order Coverage (MOC)".

doc/install.sgm
```diff
@@ -1,123 +1,98 @@
...
-<![CDATA[shell> tar -xzf path/to/pg_sphere_xxx.tgz]]>
-            </programlisting>
-            <para>
-                Now, change into the <filename>pg_sphere</filename>
-                directory and run :
-            </para>
-            <programlisting>
+  <para>
+   &pgsphere; is not the part of the &postgresql; software. You can download
+   the latest release sources from the
```
esabol, line 9: Remove "sources" from this line. Saying "latest release" is sufficient and more succint.

doc/install.sgm
```diff
@@ -1,123 +1,98 @@
...
+   <ulink url="&pgsphereurl;">&pgsphere; Releases page</ulink>.
+   The sources can also be downloaded by cloning the repository with the
```
esabol, line 11: Change "sources" to "source code". Native English speakers wouldn't use the plural in this context.

doc/install.sgm
```diff
@@ -1,123 +1,98 @@
...
+  </para>
+
+  <para>Unpack the downloaded archive and enter the directory:</para>
+  <programlisting>
+<![CDATA[shell> tar -xzf path/to/pgsphere-X.X.X.tgz]]>
+<![CDATA[shell> cd pgsphere-X.X.X]]>
+  </programlisting>
+
+  <para>
+   Compile the code. By default, &pgsphere; is compiled with the HEALPIX
```
esabol, line 40: Change "the HEALPIX" to "HEALPix" (no "the" and lowercase "ix").

doc/install.sgm
```diff
@@ -1,123 +1,98 @@
...
+  </para>
+  <programlisting>
 <![CDATA[shell> make]]>
-            </programlisting>
-            <para>
-                and to install <application>pgSphere</application> :
-            </para>
-            <programlisting>
+  </programlisting>
+  <para>or compile without HEALPIX support:</para>
```
esabol, line 46: Change "HEALPIX" to "HEALPix" here as well.

doc/install.sgm
```diff
@@ -1,123 +1,98 @@
...
+  <programlisting>
+<![CDATA[shell> make USE_HEALPIX=0]]>
+  </programlisting>
+
+  <para>
+   Run regression tests optionally. If the &pgsphere; was compiled without
+   HEALPIX support, USE_HEALPIX=0 should be specified in make command line.
```
esabol, line 53: Change "HEALPIX" to "HEALPix" here as well, but only the first instance. I recommend changing "specified in make command line" to "added after make in the following command".

doc/install.sgm
```diff
@@ -1,123 +1,98 @@
...
-            </para>
-            <programlisting>
+  </programlisting>
+  <para>or compile without HEALPIX support:</para>
+  <programlisting>
+<![CDATA[shell> make USE_HEALPIX=0]]>
+  </programlisting>
+
+  <para>
+   Run regression tests optionally. If the &pgsphere; was compiled without
```
esabol, line 52: Remove "the" before "&pgsphere;".

doc/install.sgm
```diff
@@ -1,123 +1,98 @@
...
+   HEALPIX support, USE_HEALPIX=0 should be specified in make command line.
+  </para>
+  <programlisting>
+<![CDATA[shell> make test]]>
+  </programlisting>
+
+  <para>
+   Install &pgsphere; files to the installation directories. The installation
+   directories are defined by &pg_config; utility. Superuser (root) access
+   rights may be required.
```
esabol, line 62: I would recommend adding the setence "If &pgsphere; was compiled without HEALPix support, USE_HEALPIX=0 should be added after make in the following command."

doc/install.sgm
```diff
@@ -1,123 +1,98 @@
...
-                Please have a look at the
-                <application>PostgreSQL</application> documentation for more
-                details.
-            </para>
-            <para>
-                To get the version of installed pgSphere software, simply
-                call:
-            </para>
-            <programlisting>
+  <para>To get the version of installed pgSphere software:</para>
```
esabol, line 92: Change "pgSphere" to "&pgsphere;" here?

esabol (changes requested):

doc/install.sgm
vitcpp: I thought that make install is enough and adding USE_HEALPIX=0 install is not necessary. There are some cases when developer compiles and installs the library using make install command only. In this case USE_HEALPIX should be added to disable the healpix support.

esabol: I think that is incorrect and you must specify `USE_HEALPIX=0` when you `make install`, but, if you have tried it and it works, then I'll take your word for it. I'm pretty sure I tried it once and it caused problems, but I might be remembering `make test` or another target.

vitcpp: @esabol You are right, USE_HEALPIX=0 should be added to make install. Otherwise, it will compile moc.c and process_moc.cpp. Anyway, I've modified the doc as you suggested. Thank you!

doc/install.sgm
```diff
@@ -1,123 +1,101 @@
...
+  </para>
+
+  <para>Unpack the downloaded archive and enter the directory:</para>
+  <programlisting>
+<![CDATA[shell> tar -xzf path/to/pgsphere-X.X.X.tgz]]>
+<![CDATA[shell> cd pgsphere-X.X.X]]>
+  </programlisting>
+
+  <para>
+   Compile the code. By default, &pgsphere; is compiled with the &healpix;
```
esabol, line 42: Please remove "the" before "&healpix;" on this line. (Sorry I didn't notice this in my previous review.)

esabol (changes requested):
One more change, please. Thank you!

doc/install.sgm
vitcpp: @esabol It's not a big deal. I've removed the article and rebased the branch. Thank you!

esabol (approved):
Looks good! Thank you!

vitcpp:
Rebased

esabol (approved):

PR #68, opened 2023-09-26
Fix compilation error (-Wstringop-overflow)

vitcpp:
GCC may produce false warnings in some cases when working with strings (char*) when -Wstringop-overflow is enabled. The patch fixes the compilation warning that origins in some tricky pointers casts.

The error message:

`
g++ -Wall -Wpointer-arith -Wendif-labels -Wmissing-format-attribute -Wimplicit-fallthrough=3 -Wcast-function-type -Wshadow=compatible-local -Wformat-security -fno-strict-aliasing -fwrapv -g -O2 -fPIC -fvisibility=hidden -fvisibility-inlines-hidden -DPGSPHERE_VERSION=1.3.1 -I. -I./ -I/home/vit/p.pgsphere/local/include/postgresql/server -I/home/vit/p.pgsphere/local/include/postgresql/internal  -D_GNU_SOURCE  -fopenmp -I/usr/include/healpix_cxx  -c -o src/process_moc.o src/process_moc.cpp -MMD -MP -MF .deps/process_moc.Po
In file included from /usr/include/string.h:535,
                 from /usr/include/c++/11/cstring:42,
                 from src/process_moc.cpp:2:
In function ‘void* memmove(void*, const void*, size_t)’,
    inlined from ‘void rpage_iter<V, page_size, value_size>::set(const V&) [with V = moc_interval; long unsigned int page_size = 1996; long unsigned int value_size = 16]’ at src/process_moc.cpp:336:15,
    inlined from ‘void create_moc_release_context(void*, Smoc*, pgs_error_handler)’ at src/process_moc.cpp:604:9:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:36:34: warning: writing 16 bytes into a region of size 2 [-Wstringop-overflow=]
   36 |   return __builtin___memmove_chk (__dest, __src, __len,
      |          ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
   37 |                                   __glibc_objsize0 (__dest));
      |                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from src/process_moc.cpp:19:
src/pgs_process_moc.h: In function ‘void create_moc_release_context(void*, Smoc*, pgs_error_handler)’:
src/pgs_process_moc.h:58:25: note: destination object ‘Smoc::version’ of size 2
   58 |         uint16          version;        /* version of the 'toasty' MOC data structure */
`

PR #69, opened 2023-09-26
Add GitHub Actions pipeline for pgSphere

borodun:
Use GitHub Actions for pgSphere pipeline. At this moment, builds with HEALPix are failing due to compilation warning, but https://github.com/postgrespro/pgsphere/pull/68 should fix this.

### DIFF
--- a/.github/workflows/build-and-check.yml
+++ b/.github/workflows/build-and-check.yml
@@ -1,0 +1,67 @@
+name: Build and Check
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  build_and_test:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        pg_version: [10, 11, 12, 13, 14, 15, 16]
+        use_healpix: [0, 1]
+
+    name: PostgreSQL ${{ matrix.pg_version }} - USE_HEALPIX=${{ matrix.use_healpix }}
+
+    steps:
+    - name: Install dependencies
+      run: |
+        sudo apt update && sudo apt install -y \
+          postgresql-common \
+          libhealpix-cxx-dev \
+          docbook-xml \
+          docbook-xsl \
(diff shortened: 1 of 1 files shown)


esabol (approved):
I love it! Fantastic contribution!

esabol:
I also dream of a GitHub Actions pipeline which, on merges to the  master branch, uploads the latest documentation to the GitHub Pages associated with this repository.

borodun:
@esabol, I moved `make test` before `make install`. About `fail-fast: false`: I wanted to make the CI behavior the same as the current Travis pipeline, it will not cancel jobs if other jobs fail first. In my opinion, this will help to get more informative pipeline runs.

borodun:
@esabol, regarding the pipeline for docs: that’s the plan. We need to organize access to GitHub Pages for this repository. This is a more administrative issue. We want to get this done as soon as possible.

esabol:
The new GitHub Actions pipeline doesn't seem to be working in PR #65, unfortunately. Very weird error that I've not encountered before in my GitHub Actions.

vitcpp:
@esabol, @borodun  It seems the problem is in line 63 of build-and-check.yml: github.ref_name is 65/merge. It seems the / symbol should be replaced with some other acceptable symbol or another var should be used.

esabol:
Changing the '/' to a '-' in `$github_ref_name` can be done as shown here:

https://github.com/mbiagetti/github-action-poc/pull/2/files

But that method seems *very* convoluted and annoying. Would this method work?

https://stackoverflow.com/a/67290085/5153779

borodun:
We can use something like https://github.com/marketplace/actions/github-slug-action to create slugged versions of variables. We can add:
```
- name: Inject slug/short variables
  uses: rlespinasse/github-slug-action@v4
```
And change `${{ github.ref_name }}` to `${{ env.GITHUB_REF_SLUG_URL }}`

I'll create PR

PR #71, opened 2023-09-28
Add "pgindent" rule for make to run pgindent

vitcpp:
Implemented pgindent rule to run pgindent on the code. Utilities pgindent and pg_bsd_indent should be in the PATH or specified in the make command line: make PGINDENT=<...> PGBSDINDENT=<...> pgindent


esabol (approved):

pgindent-typedefs.list
```diff
@@ -0,0 +1,236 @@
...
+wint_t
+YY_BUFFER_STATE
+YY_CHAR
+yy_size_t
+yy_state_type
+YYSTYPE
+yytype_int16
+yytype_int8
+yytype_uint8
+CPoint
```
df7cb, line 236: I'd suggest to sort the list - it looks like it is, except for that last value.

esabol:
Where do you get `pgbsdindent`? My `postgresql-15.4/src/tools/pgindent/` directory only compiles the `pgindent` and `pgperltidy` executables....

Hmmm, the `README` in `postgresql-15.4/src/tools/pgindent/` says to `git clone https://git.postgresql.org/git/pg_bsd_indent.git`. That compiles `pg_bsd_indent`, not `pgbsdindent`. I'm confused as to the difference. I presume they are compatible, but why is the name of the executable different?

Might be time to add a `CONTRIBUTING.md` file to the repo to explain this stuff?

df7cb:
pg_bsd_indent got imported into the postgresql git, but only in the current master branch. The idea is that different branches might have different indentation rules or something.

https://git.postgresql.org/gitweb/?p=postgresql.git;a=tree;f=src/tools/pg_bsd_indent

(`pg_bsd_indent` is the correct spelling, `pgbsdindent` doesn't exist. But `pgindent` exists, for added confusion...)

df7cb:
At the top of the pgindent/README file, there is a link to a blog post that explains how to use it on non-core code:

https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/tools/pgindent/README

http://adpgtech.blogspot.com/2015/05/running-pgindent-on-non-core-code-or.html

Dropping a link to at least the first one would make sense, perhaps as a comment on the Makefile pgindent rule.

esabol:
Oops, I'm not sure why I thought the executable is named "pgbsdindent". The PR description and the changes to the `Makefile` clearly show its "pg_bsd_indent". I guess I was just temporarily blind. Anyway, thanks for clarifying, @df7cb.

I guess I'd still like to see some text added to the `README.pg_sphere` file **or** comment(s) added to the `Makefile` **or** a `CONTRIBUTING.md` file added to the repo which explains where/how to get the `pgindent` and `pg_bsd_indent` executables.

vitcpp:
@esabol I think, we have to put some explanations to CONTRIBUTING.md but I propose to do it later. Now CONTRIBUTING.md is too old and not complete. It should be updated. What I propose to do right now is to put some information into the Makefile (above the rule). To run pgindent a postgresql source tree should be installed because it contains sources of pg_bsd_indent and pgindent, or these files should be somehow available in PATH by some other means. The pgindent, pg_bsd_indent binaries should be in the PATH. 

I agree with @df7cb to put the link into the Makefile:  
https://git.postgresql.org/gitweb/p=postgresql.git;a=blob;f=src/tools/pgindent/README

df7cb:
Fwiw I plan to change the PG17 Debian packages to include pg_bsd_indent and pgindent, if that helps.

https://wiki.postgresql.org/wiki/Apt/FAQ#Development_snapshots

Makefile
```diff
@@ -223,3 +225,27 @@ endif
 dist : clean sparse.c sscan.c
 	find . -name '*~' -type f -exec rm {} \;
 	cd .. && tar --transform s/$(SRC_DIR)/pgsphere-$(PGSPHERE_VERSION)/ --exclude CVS --exclude .git -czf pgsphere-$(PGSPHERE_VERSION).tar.gz $(SRC_DIR) && cd -
+
+# To launch pgindent, set the PATH environment variable to the directories
```
esabol, line 229: I propose changing this line to the following:

To use pgindent, set the PATH environment variable to include the directories

Makefile
```diff
@@ -223,3 +225,27 @@ endif
...
+# containing the binaries pgindent and pg_bsd_indent. It is important to
+# utilize a specific version of pg_bsd_indent, which sources can be found
+# in the postgresql>/src/tools/pg_bsd_indent directory, where <postgresql>
```
esabol, line 232: There should be a less-than symbol before "postgresql" on this line, I believe.

esabol (changes requested):

vitcpp:
Rebased, fixed issues reported in the review

esabol (commented):
Looks good!

PR #72, opened 2023-09-28
Format the code with pgindent

vitcpp:


esabol:
I don't really like pgindent's excessive whitespace, but I'm ok with it if this is something you want.

vitcpp:
@esabol I would suggest to consider this PR as a proposal, not a solution to be applied unconditionally. I agree with you about excessive indentation and I'm not a fan pgindent. Furthermore, the formatting of some mathematical expressions should be untouched. But I do not know another solution that may help us with the indentation. We may use pgindent as a code formatting verification tool. We will be able to automate this process. Anyway, I'm ok if we decide to reject this PR and do not use pgindent.

esabol:
I think it’s up to you. I don’t have a strong opinion on it.

vitcpp:
I think it would be better to limit the pgindent functionality somehow. I would propose to fix only identations at the beginnings of the strings, and function signatures. The formatting of formulae should be untouched.

vitcpp:
Rejected, because it is outdated.

PR #73, opened 2023-09-28
Add pipeline for docs using GitHub Pages

borodun:
Build and deploy docs to GitHub Pages. This workflow will deploy for master branch. For now, it hosts doc/html directory, so only the last version will be available. How necessary is it to organize the deployment of documentation with previous versions?

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -1,0 +1,51 @@
+name: Build and Deploy Docs
+
+# Deploy docs only for master
+on:
+  push:
+    branches:
+      - "master"
+
+# Allow deployment to GitHub Pages
+permissions:
+  pages: write
+  id-token: write
+
+jobs:
+  deploy-docs:
+
+    name: Deploy Docs
+
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Install dependencies
+      run: |
+        sudo apt update && sudo apt install -y \
+          docbook-xml \
+          docbook-xsl \
+          libxml2-utils \
+          xsltproc \
+          fop
+
+    - name: Clone pgSphere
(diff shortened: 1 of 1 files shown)


esabol (approved):
This is great! Thank you!

I think we only care about the latest documentation?

Is there a way to test this before we merge the PR? Otherwise, we can always tweak after merging.

borodun:
@esabol, I've tested it on my fork by removing `branches: master`: https://github.com/borodun/pgsphere/actions/runs/6339680394. Docs were deployed to https://borodun.github.io/pgsphere/.

esabol (approved):
Looks good! Thank you!

esabol:
So what's the URL of the GitHub Pages documentation? We should add this URL to the README and possibly the documentation, I think.

I tried https://postgrespro.github.io/pgsphere/ and got a 404 error, so that must not be it.

borodun:
@esabol, @vitcpp It looks like the following job https://github.com/postgrespro/pgsphere/actions/runs/6403043991 (default Jekyll workflow) was executed after https://github.com/postgrespro/pgsphere/actions/runs/6403044188 and erased all static files. After re-run, docs appeared again on https://postgrespro.github.io/pgsphere/. We need to disable this workflow: https://github.com/postgrespro/pgsphere/actions/workflows/pages/pages-build-deployment. Seems like it was added due to mistake in repo setting, because it doesn't have a corresponding workflow file in `.github/workflows` directory.

esabol:
I was reading [the documentation on this](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site) yesterday, and it said something about creating a `.nojekyll` file, and I didn't understand why. Maybe this is the reason?

borodun:
@esabol, thank you! As I understand from the docs, this file is only used when publishing Github Pages from a branch. We have tried to change repo settings to use Github Actions to publish Pages. We will see if this helped when new commits will be added to master. So, for now, this file won't be necessary, I think.

PR #74, opened 2023-10-02
Spoint3 fetch

df7cb:
Next step in cleaning up the Makefile: Make support for index only scans in the spoint3 gist opclass official


esabol (approved):
Looks good to me!

vitcpp (approved):

PR #78, opened 2023-10-05
Use a static filename for the header of RELEASE_SQL

df7cb:
One place less to update when bumping the version number.


esabol (approved):
Very sensible! Thank you!

vitcpp (approved):

PR #79, opened 2023-10-05
make test: Don't track NOTICEs

df7cb:
There's little value in tracking the noise of NOTICEs emitted during loading pg_sphere.test.sql, so just mute them.

The gain from this change is that future patches will not have to update the changed line numbers in the output tracked. (Cf. c2a235160)

`\set ECHO all` at the end of the file can be removed since a new psql session will be created for the next file anyway.


esabol (approved):
This addresses something that I (and other contributors, I think) have found to be very painful. Thank you!

vitcpp:
Cool. Thank you!

df7cb:
Rebased to master.

PR #83, opened 2023-10-16
g_spoint3_fetch was already part of the spoint3 opclass in older vers…

### LINKED ISSUE #82: g_spoint3_fetch wasn't "too experimental" before 1.1.5
As just mentioned in #75, there is a problem with the spoint3 change from #74:

In version 1.0, the opclass already contained the function `g_spoint3_fetch` that #74 is adding to 1.3.2.
The function was declared "too experimental and preliminary" in bdc37d186f which is a mega-commit that labelled "pgsphere_1.1.4.916", so later installs don't contain it, but older installations upgraded since then will already have it.

Do we care? The fix would likely be wrapping the "alter operator family" call in upgrade_scripts/pg_sphere--1.3.1--1.3.2.sql.in into a DO block that would ignore the error when the function is already present.

df7cb:
…ions

Allow upgrading of older installations (made pre 1.1.5) that already have g_spoint3_fetch in the spoint3 opclass.

Close #82.

### DIFF
--- a/upgrade_scripts/pg_sphere--1.3.1--1.3.2.sql.in
+++ b/upgrade_scripts/pg_sphere--1.3.1--1.3.2.sql.in
@@ -1,4 +1,14 @@
 -- add "fetch" support function to enable index-only scans for spoint3
 
-ALTER OPERATOR FAMILY spoint3 USING gist ADD
-   FUNCTION 9 (spoint, spoint) g_spoint3_fetch (internal);
+-- g_spoint3_fetch was already part of the spoint3 opclass in older versions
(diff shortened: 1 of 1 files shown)


esabol (approved):

vitcpp (approved):

esabol:
@vitcpp I propose releasing version 1.3.2.

df7cb:
Or perhaps 1.4.0?

vitcpp:
Yes, 1.4.0. I think, we do it in a separate PR as a final PR before the new release. I will create an issue for it.

PR #84, opened 2023-10-23
Maintain only one list of tests

df7cb:
Make evaluates "=" variables lazily, so we can add items to TESTS later in the file and still have them show up in REGRESS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,13 +33,10 @@ DATA_built  = $(RELEASE_SQL) \
 			  pg_sphere--1.3.1--1.3.2.sql
 
 DOCS        = README.pg_sphere COPYRIGHT.pg_sphere
-REGRESS     = init tables points euler circle line ellipse poly path box index \
-              contains_ops contains_ops_compat bounding_box_gist gnomo epochprop \
-              contains overlaps spoint_brin sbox_brin selectivity
-
-TESTS       = init_test tables points euler circle line ellipse poly path box \
+TESTS       = tables points euler circle line ellipse poly path box \
               index contains_ops contains_ops_compat bounding_box_gist gnomo \
               epochprop contains overlaps spoint_brin sbox_brin selectivity
+REGRESS     = init $(TESTS)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -104,17 +100,17 @@ healpix_bare/healpix_bare.o : healpix_bare/healpix_bare.c
 pg_version := $(word 2,$(shell $(PG_CONFIG) --version))
 has_support_functions = $(if $(filter-out 9.% 10.% 11.%,$(pg_version)),y,n)
 
-crushtest: REGRESS += $(CRUSH_TESTS)
+crushtest: TESTS += $(CRUSH_TESTS)
```
esabol, line 103: This change leaves out `init` from `crushtest`, right? Why would we want to do that?

df7cb: crushtest is unaffected by this change, it runs the normal "installcheck" target, just with more tests. The list of tests is in $(TESTS) now, $(REGRESS) is only used internally by the pgxs rules that invoke pg_regress.

Adding to REGRESS would also work for this use case, but all the other rules in the Makefile need to add to TESTS so "make test" is also supported, so I think it's cleaner to have crushtest append to TESTS as well.

esabol: Sorry, where is the `installcheck` target defined? I can't find it.

df7cb: In PostgreSQL's pgxs.mk. The tests are run on `$(REGRESS)`, our Makefile defines `REGRESS     = init $(TESTS)`, crushtest appends to `TESTS`, so when `crushtest` is run, it includes the crush files.

esabol: Well, OK then!

esabol (approved):

vitcpp (approved):

df7cb:
@vitcpp I can rebase the PRs in whatever order you want to merge

vitcpp:
@df7cb ok, thanks. Lets start from the simpler PRs then.

PR #86, opened 2023-10-23
Configurable smoc_gin_ops index resolution

df7cb:
Back when smoc GIN indexes were implemented, indexes did not support opclass options yet; that was added only later in PG 13.

Add an "order" parameter on the smoc_gin_ops opclass to allow picking the smoc index granularity between level 0 and 12. (Larger levels do not fit into the internal int32 datatype anymore.)

Example:
`create index on sky using gin (coverage smoc_gin_ops (order = 8))`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,6 +99,7 @@ healpix_bare/healpix_bare.o : healpix_bare/healpix_bare.c
 
 pg_version := $(word 2,$(shell $(PG_CONFIG) --version))
 has_support_functions = $(if $(filter-out 9.% 10.% 11.%,$(pg_version)),y,n)
+has_index_options = $(if $(filter-out 9.% 10.% 11.% 12.%,$(pg_version)),y,n)
 
 crushtest: TESTS += $(CRUSH_TESTS)
 crushtest: installcheck
@@ -106,6 +107,13 @@ crushtest: installcheck
 ifeq ($(has_support_functions),y)
 PGS_SQL    += pgs_gist_support.sql
 TESTS      += gist_support
+endif
+
+ifneq ($(USE_HEALPIX),0)
+ifeq ($(has_index_options),y)
+PGS_SQL    += pgs_moc_options.sql
+TESTS      += moc_options
+endif
 endif
 
 # "make test" uses a special initialization file that doesn't rely on "create extension"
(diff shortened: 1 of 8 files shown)


msdemlei (approved):
This is a useful new feature for me, so +1 on that account.  I've lightly reviewed the code and have found no reason to complain (disclaimer: I'm no expert for PG indexing).

doc/indices.sgm
```diff
@@ -121,23 +121,32 @@
...
           <programlisting>
 <![CDATA[CREATE TABLE ivoa (]]>
 <![CDATA[  coverage smoc NOT NULL]]>
 <![CDATA[);]]>
 <![CDATA[-- Put in data now]]>
+<![CDATA[-- Create index with the defaut smoc_gin_ops opclass]]>
 <![CDATA[CREATE INDEX ON ivoa USING GIN (coverage);]]>
 <![CDATA[-- Alternative index with more detail]]>
+<![CDATA[CREATE INDEX ivoa_order_7_idx ON ivoa USING GIN (coverage smoc_gin_ops (order = 7));]]>
+<![CDATA[-- Alternative operator class with fixed order 8]]>
```
vitcpp, line 149: Not sure, order = **7**, but the text contains 'with fixed order **8**'. Is it correct?

vitcpp (changes requested):

doc/indices.sgm
df7cb: It it correct, the 7 belongs to "more detail" on the line before, and order 8 is for the _fine opclass on the next line. But you are right, it is confusing, I'll make it more explicit.

Makefile
```diff
@@ -186,6 +194,11 @@ pg_sphere--1.3.0--1.3.1.sql:
 ifeq ($(has_support_functions),y)
 pg_sphere--1.3.1--1.3.2.sql: pgs_gist_support.sql.in
 endif
+ifneq ($(USE_HEALPIX),0)
+ifeq ($(has_index_options),y)
+pg_sphere--1.3.1--1.3.2.sql: pgs_moc_options.sql.in
+endif
+endif
 pg_sphere--1.3.1--1.3.2.sql: pgs_circle_sel.sql.in
```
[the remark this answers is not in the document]
df7cb: TBH I'd rather not to keep all the code for `pg_sphere--1.3.1--1.3.2.sql` in one code block. It does look ugly, but I think keeping it together makes it immediately clear that it's one logical block.

vitcpp (approved):

PR #87, opened 2023-10-23
Use explain() function in moc100 test

df7cb:
The interesting part of the moc100 test are the row counts and "Rows Removed by Filter". Remove all other noise. This also lets us get rid of the moc100_N alternative output files.


msdemlei (approved):
Thanks!

vitcpp (approved):

PR #92, opened 2023-10-26
Drop obsolete compatibility @({bigint,spoint}, smoc) operators

### LINKED ISSUE #85: Operator `@` is incompatible with PG17
PG17 doesn't like the `bigint @ smoc` and `smoc @ bigint` operators:

```
+psql:pg_sphere.test.sql:9968: ERROR:  commutator operator @> is already the commutator of operator <@
+psql:pg_sphere.test.sql:9978: ERROR:  commutator operator @> is already the commutator of operator <@
```

This is from pgs_moc_compat.sql.in:

```
-- a minimal set of backwards-compatible operators for GAVO DaCHS,
-- see http://soft.g-vo.org/dachs

CREATE OPERATOR @ (
   LEFTARG    = bigint,
   RIGHTARG   = smoc,
   PROCEDURE  = healpix_subset_smoc,
   COMMUTATOR = '@>',
   NEGATOR    = '!<@',
   RESTRICT   = contsel,
   JOIN       = contjoinsel
);

CREATE OPERATOR @ (
   LEFTARG    = spoint,
   RIGHTARG   = smoc,
   PROCEDURE  = spoint_subset_smoc,
   COMMUTATOR = '@>',
   NEGATOR    = '!<@',
   RESTRICT   = contsel,
   JOIN       = contjoinsel
);
```

The IMHO best fix would be to drop the `@` operator. Alternatively, we can drop the COMMUTATOR and NEGATOR clauses.

Comments? 

Cc: @msdemlei

df7cb:
These were required by GAVO DaCHS, but are not used anymore. Furthermore, PG17 will forbid declaring commutators and negators relations with operators that themselves have relations to different operators.

We delete pgs_moc_compat.sql.in and remove it from the old pg_sphere--1.1.5beta2gavo--1.1.5beta4gavo.sql upgrade script. This is ok to do since dropping the operators uses "if exists".

Discussion: https://www.postgresql.org/message-id/flat/ZTfL1G1fBPQHVPWV%40msg.df7cb.de#ce3801c084bcf05b6b976cb2d0e25b2d

Close #85.


df7cb:
Sorry, I had typoed the "git push --force" command and mistakenly deleted the branch which auto-closed #91. This is the same patch rebased to current master.

esabol (approved):
LGTM.

vitcpp (approved):

vitcpp:
@df7cb Thank you for the PRs! I propose to make the new release at the beginning of the next week if no new PRs will appear.

df7cb:
I have another indexing idea in the pipeline that would make sense including with 1.4, so please wait a bit. I should be able to submit it this week.

Issue #95, opened 2023-10-27
Additional spoly constructors?


esabol:
The `spoly_deg()` function can construct a `spoly` from an array of `float8`:
```
CREATE FUNCTION spoly_deg(float8[])
   RETURNS spoly
   AS 'MODULE_PATHNAME', 'spherepoly_deg'
   LANGUAGE 'c'
   IMMUTABLE STRICT;
```
(Aside: Why is this function not PARALLEL SAFE? Should it be?)

But there's no corresponding `spoly` constructor that takes an array of `float8` in radians:
```
testdb=> SELECT spoly(ARRAY[1.0, 2.0, 3.0, 4.0, 5.0, 6.0]);
ERROR:  function spoly(numeric[]) does not exist
LINE 1: SELECT spoly(ARRAY[1.0, 2.0, 3.0, 4.0, 5.0, 6.0]);
               ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```
I don't want to use a temporary table to construct a `spoly` in my PLpgsql functions. It's horribly inefficient and doesn't scale when operating over very large tables. So that means I have to either convert my coordinates to degrees (only for `spherepoly_deg()` to convert them back to radians) in order to avoid that or construct a string to contain the array of coordinates and use the `spoly` constructor on the string.

Also, I think it would be nice if there was a constructor that takes an array of `spoint` as well, so that you can do this:
```
SELECT spoly([spoint(1.0, 2.0), spoint(3.0, 4.0), spoint(5.0, 6.0)]);
```
Should that be possible? What do you think?

vitcpp:
@esabol I think, it is possible what you proposes. Postgresql supports function overloading.

esabol:
How would you advise implementing the `spoly(spoint[])` constructor function?

Would the spoint array need to be duplicated in memory? Can you reuse the same array and/or same array members passed in as arguments?

esabol:
It feels like there should be some trick to creating an `spoly` from an array of points since you can sort of already do it like this:
```SQL
select spoly(t.pnt) into thepoly 
	from (select pnt from xxxpnts) as t;
```

vitcpp:
@esabol I think, this issue is completed by the following PRs:
- #96 
- #99

PR #96, opened 2023-10-28
Add function to create spoly from an array of numbers in radians

esabol:
As discussed in issue #95, this pull request adds a constructor function to create an spoly from an array of numbers in radians. It is very similar to spoly_deg() except without the calls to `deg_to_rad()`.

Question 1: Can the spoly(float8[]) function be marked PARALLEL SAFE? My reading of the documentation for PARALLEL SAFE doesn't mention anything about accessing arrays or allocating memory, so I think it should be.

Question 2: Can it be marked STRICT? `select spoly(NULL::float8[])` returns null, I believe.

If the answer to either or both questions, I would like to make the same change(s) to `spoly_deg` also.

### DIFF
--- a/expected/poly.out
+++ b/expected/poly.out
@@ -318,6 +318,18 @@ SELECT spoly '{(10d,0d),(10d,1d),(15d,0d)}';
  {(10d , 0d),(10d , 1d),(15d , 0d)}
 (1 row)
 
+SELECT spoly(ARRAY[0.017453292519943295, 0.03490658503988659, 0.05235987755982988, 0.06981317007977318, 0.08726646259971647, 0.10471975511965977]);
+              spoly              
+---------------------------------
+ {(1d , 2d),(3d , 4d),(5d , 6d)}
+(1 row)
+
+SELECT spoly(ARRAY[0.17453292519943295, 0.0, 0.17453292519943295, 0.017453292519943295, 0.2617993877991494, 0.0]);
+               spoly                
+------------------------------------
+ {(10d , 0d),(10d , 1d),(15d , 0d)}
+(1 row)
+
 SELECT spoly_deg(ARRAY[1.0, 2.0, 3.0, 4.0, 5.0, 6.0]);
             spoly_deg            
 ---------------------------------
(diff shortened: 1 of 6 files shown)


df7cb:
I believe the answer to both question is "yes".

Mind that besides the parameter being NULL, individual array elements might also be NULL.

src/polygon.c
```diff
@@ -906,6 +907,51 @@ spherepoly_in(PG_FUNCTION_ARGS)
...
+			);
+		PG_RETURN_NULL();
+	}
+
+	array_data = (float8 *) ARR_DATA_PTR(float_vector);
+
+	for (i = 0; i < np; i++)
+	{
+		create_spherepoint_from_long_lat(
+										 &points[i],
```
vitcpp, line 947: Move &points... to the line above.

src/polygon.c
```diff
@@ -906,6 +907,51 @@ spherepoly_in(PG_FUNCTION_ARGS)
 	PG_RETURN_POINTER(poly);
 }
 
+Datum
+spherepoly_rad(PG_FUNCTION_ARGS)
```
vitcpp, line 911: I think, in this implementation we do not properly process arrays with some NULL elements. There is a number of functions like ARR_HASNULL, ARR_NULLBITMAP that may help to identify NULL array elements. The simple fix is to reject such arrays with NULL elements.

It seems for spherepoly_deg we missed the check for NULL array elements as well.

I'm ok to mark it as PARALLEL SAFE and STRICT. I think we mssed to mark spherepoly_deg because it was under development when we were adding PARALLEL SAFE and STRICT.

vitcpp (changes requested):

src/polygon.c
esabol: @vitcpp : I think I have addressed all of your review items. Please check. I have left the commits unsquashed to facilitate your review. I will rebase and squash the commits before merging.

expected/poly.out
```diff
@@ -349,6 +349,8 @@ LINE 1: SELECT spoly '{(10d,0d),(10d,1d)}';
                      ^
 SELECT spoly(ARRAY[1.0, 2.0, 3.0, 4.0, 5.0]);
 ERROR:  spherepoly_rad: invalid number of arguments (must be even and >= 6)
+SELECT spoly(ARRAY[1.0, 2.0, 3.0, NULL, 5.0, 6.0]);
+ERROR:  spherepoly_rad: invalid array has null values
```
vitcpp, line 353: Not sure, that it is a good sentence. May be something like it: The input array is invalid because it contains null values? Anyway, I'm not a native speaker. It is up to you.

vitcpp (changes requested):

vitcpp:
@esabol I think, everything ok except of some trivial formatting changes. Thank you!

src/polygon.c
```diff
@@ -920,8 +920,7 @@ spherepoly_rad(PG_FUNCTION_ARGS)
 
 	if (np < 6 || np % 2 != 0)
 	{
-		elog(
-			 ERROR,
+		elog(	 ERROR,
```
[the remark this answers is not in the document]
esabol: OK, how about now?

expected/poly.out
esabol: I think my error message was OK, but your version is clearer, so I have changed it to your version.

vitcpp:
@esabol I'm ok to accept the patch. Could you please squash the commits?

esabol:
Rebased and squashed. Ready to merge whenever you are. Thank you for your reviews, @vitcpp !

vitcpp (approved):

vitcpp:
It seems I'm the only reviewer. Lets merge it, since the risks to break the pgpshere are minimal.

Issue #112, opened 2023-12-01
spoly that was valid now detected as invalid


pdowler:
I have a polygon from an existing server that later versions of pgsphere detect as invalid. 

Summary: The polygon below is acceptable in pgsphere up to 1.2.3 and then rejected by 1.3.1 and 1.4.1.

```
psql (15.4, server 15.5)
Type "help" for help.

content=> select pg_sphere_version();
 pg_sphere_version 
-------------------
 1.2.3
(1 row)

content=> select '{
(3.30474723646012 , 1.08600456205300),
(3.30341855309927 , 1.08577960186707),
(3.30341054542378 , 1.08578643990271),
(3.30297351563319 , 1.08633534556428),
(3.30357156120003 , 1.08643683957210),
(3.30358891855857 , 1.08643995044436),
(3.30360894676365 , 1.08644306147078),
(3.30361829343581 , 1.08644430596871),
(3.30362630482521 , 1.08644555030213),
(3.30364633346451 , 1.08644866102000),
(3.30365300940335 , 1.08645052692055),
(3.30366102096957 , 1.08645177113937),
(3.30367036769496 , 1.08645363721023),
(3.30367837934959 , 1.08645488137174),
(3.30368906174976 , 1.08645612569695),
(3.30370107936906 , 1.08645799183673),
(3.30370642025712 , 1.08645985750225),
(3.30373179124734 , 1.08646358962156),
(3.30374514456618 , 1.08646545561358),
(3.30410706158729 , 1.08652886672786),
(3.30427803417922 , 1.08655868846497),
(3.30429673329093 , 1.08655930694968),
(3.30432478121775 , 1.08655930174652),
(3.30433278932944 , 1.08655308246640),
(3.30446348355532 , 1.08638330933224)
}'::spoly;

spoly                                                                                             
--------
 {(3.30474723646012 , 1.086004562053),(3.30341855309927 , 1.08577960186707),(3.30341054542378 , 1.08578643990271),(3.30297351563319 , 1.086335
34556428),(3.30357156120003 , 1.0864368395721),(3.30358891855857 , 1.08643995044436),(3.30360894676365 , 1.08644306147078),(3.30361829343581 ,
 1.08644430596871),(3.30362630482521 , 1.08644555030213),(3.30364633346451 , 1.08644866102),(3.30365300940335 , 1.08645052692055),(3.303661020
[...]

PR #113, opened 2023-12-05
Fix polygon validation function (issue #112)

vitcpp:


expected/poly.out
```diff
@@ -403,13 +403,17 @@ LINE 1: SELECT spoly '{(0d,0d),(10d,10d),(0d,10d),(10d,0d)}';
...
 --- degenerate polygons -----
 SELECT spoly '{(0d,1d),(0d,2d),(0d,3d)}';
-ERROR:  spherepoly_from_array: a line segment overlaps or polygon too large
-LINE 1: SELECT spoly '{(0d,1d),(0d,2d),(0d,3d)}';
-                     ^
+              spoly              
+---------------------------------
+ {(0d , 1d),(0d , 2d),(0d , 3d)}
+(1 row)
+
```
esabol, line 410: This doesn't seem correct. This spoly and the following one _should_ fail, no? They do not seem to be valid polygons to me.

vitcpp: @esabol I rolled back the changes in PR #36 because it broke the creation of valid polygons. I looked at the issue where we decided to consider degenerate polygons as invalid, but I can't remember why we decided to so, unfortunately. Probably, due to OGC specification of valid polygons. This bug actually makes pgsphere unusable in production, but invalidation of degenerate polygons don't produce any problems. Despite of OGC specification of valid polygons, in practice, we may consider support of genenerate polygons as well to make pgsphere more usable.

src/line.c
```diff
@@ -496,6 +496,15 @@ sline_sline_pos(const SLine *l1, const SLine *l2)
...
 
+	/* check connected lines */
+	if (FPgt(il2->length, 0.0) && (vector3d_eq(&v[0][0], &v[1][0]) ||
+								   vector3d_eq(&v[0][0], &v[1][1]) ||
+								   vector3d_eq(&v[0][1], &v[1][0]) ||
+								   vector3d_eq(&v[0][1], &v[1][1])))
+	{
+		return PGS_LINE_CONNECT;
+	}
+
```
esabol, line 507: Would it be possible to add a test for this change to `sline_sline_pos()`?

vitcpp: @esabol I think, the added polygon allows to test this branch in the code. Otherwise, it goes down in the function and returns wrong relationship. The current change just rolls back the old commit. I propose to fix degenerate polygons later, in another commit.

esabol (approved):

expected/poly.out
vitcpp: @esabol I understand the problem with sline_sline_pos. I'm going to propose another solution but it takes more time to implement and test. The core reason is in using of spoint_at_sline in sline_sline_pos before checking for connectivity of adjacent segments. In past, the connectivity check was prior to the calling of spoint_at_sline.

I believe it is reasonable to apply this PR and increment patch number, because the big is severe. But next commit we improve sline_sline_pos. It more risky and needs more testing.

src/line.c
esabol: OK. Sounds good, @vitcpp.

PR #114, opened 2023-12-06
fix Makefile dist target

pdowler:
the sparse.c and sscan.c files are now committed in src dir

I use the dist target to create tarballs as part of building rpm packages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -226,7 +226,7 @@ else
 	@$(missing) flex $< $@
 endif
 
-dist : clean sparse.c sscan.c
+dist : clean
 	find . -name '*~' -type f -exec rm {} \;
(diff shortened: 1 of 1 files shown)


esabol (approved):
This isn't a Makefile target that I use, so I trust you know what's best here.

vitcpp:
@pdowler There is intention to remove these files from the src directory and auto-generate it. Can such possible change break your rule?

pdowler:
When I added that rule back in 1.1.x those files were generated in the base dir and it was the quick and dirty way to make a src tarball... wasn't the right choice back then. Now, it looks like the C files are in src and committed to git so that is no longer needed.

If there is a change in future to not have those generated C files committed, then next time around I'll fix the rpm build setup instead so it can generate them correctly.

I don't know the pros and cons of those files being committed to git... probably wouldn't have been my choice (minimalist)., but I don't intend to change that.

vitcpp (approved):

PR #117, opened 2023-12-19
Fix indentation in headers with pgindent

vitcpp:


esabol (approved):
I'm not a fan of adding `extern` to all the function prototypes in the header files (function prototypes are already implicitly `extern` in header files), but there's nothing incorrect about it, obviously. It's just a style preference thing and, if that's your preference, it's fine with me.

vitcpp:
@esabol pgindent requires the extern specifier in the beginning of the function declaration. Otherwise, it reformats function declarations in a wrong way. Once this PR is applied, pgindent will not produce new changes for headers.

Issue #118, opened 2024-01-03
Sphere output precision limited to 15 significant digits


toniorte:
Dear All,

I was wondering if there is any reason to keep the maximum precision of the printed outputs to DBL_DIG macro (15) instead of DBL_DECIMAL_DIG (17), being that just 15 digits don't allow round-trip conversion.

In my case it causes me trouble because I cannot restore dumps of databases done with pg_dump, since the dumps contain spolys which lost precision during the dump, and pg_sphere does not pass the check of certain polygons which were inserted with 17 digits (from caom2db software) but now have 15.

Regards.

esabol:
> [...] 15 digits don't allow round-trip conversion.

Oh, that's a good point.

toniorte:
Dear All,

I have tried to create a binary backup using the [pg_dumpbinary](https://github.com/lzlabs/pg_dumpbinary) software, hoping that it would keep the original data with the needed precision, but the following error appears:

**ERROR:  no binary output function available for type public.spoly**

Thus, going back to the quick solution for the text output, I guess that in fact the method _set_sphere_output_precision(n)_ (which is limited to n<=DBL_DIG, i.e. 15) tries to imitate the PostgreSQL behavior of "SET extra_float_digits".
Of course and by default, the PostgreSQL "extra_float_digits" its set to return 17 digits. And in fact, PostgreSQL documentation says that the maximum value is "useful for dumping float data that needs to be restored exactly."


Regards.

esabol:
@df7cb @vitcpp : So should it be changed to 17 on PG 12+?

vitcpp:
Dear All,

Obviously, there should be a way to set 17 digits for output. But I'm not sure about DBL_DECIMAL_DIG. It seems to appear in C11. May be to utilize extra_float_digits? Let me, please, dig into this topic. I will come with some proposal later.

df7cb:
Some other datatypes (like my `postgresql-unit`) use `extra_float_digits` for their custom float rendering on output. pgsphere should do the same.

Perhaps it is just a matter of using `float8out_internal` for printing the numbers, so something along that line.

vitcpp:
I agree, we should unify this behaviour with postgresql. What I'm concerned is the backbard compatibility. I propose the following:

- Keep function set_sphere_output_precision for some time. Probably, issue some warning that this function is obsolete.
- By default, use postgresql settings (extra_float_digits).
[...]

PR #119, opened 2024-03-14
Fix output precision limit for double values (issue #118)

vitcpp:
pgSphere used its own setting (set_sphere_output_precision function) to limit the output precision of double values, that could not be greater than 15 significant digits (DBL_DIG). It introduced some problems with dump/restore. PostgreSQL uses its own precision setting: extra_float_digits. The PostgreSQL setting allows to use more significant digits.

This patch changes the default pgSphere output behaviour to utilize PostgreSQL extra_float_digits. Now, extra_float_digits is used by default, until set_sphere_output_precision is called.

The old behaviour is kept for compatibility purposes. Once, set_sphere_output_precision is called, pgSphere starts to use the old behaviour (read, please, issue #118 discussion).

The patch introduces a new function - reset_sphere_output_precision. It is used to reset to the PostgreSQL behaviour after using set_sphere_output_precision.


vitcpp:
Hi @esabol, once I've changed the default behaviour of the output by improving the precision, some tests show differences in numbers. It is why I've added two new test variants. But, it seems it may be a mistake. It would be better to just replace the old files. I will check. Thank you. BTW, I think to add some a new test for testing output precision with different settings.

esabol:
The changes should definitely have tests. I was just confused as to how the new `expected/*.out` files did that. Variant comparison files are not something I was familiar with. Thank you for the link explaining the concept. If you think this is the best way to implement these tests, I am fine with it.

vitcpp:
expected/epochprop.out differs for 10, 11 and for 12+ versions, because the default precision was 15 significant digits prior to PG12. It is why I've added one more variant comparison file.

image

vitcpp:
> Why not put `set extra_float_digits = 2;` at the top of each "general" test file, and only check for the varying precision in a specific test for that purpose, if at all? (Ok, I see it's only the gist bounding box test that's affected, but that seems unrelated functionality.)

It makes sense, thank you! It seems, set_sphere_output_precision(8) is set at the top of almost all tests.

esabol (approved):
Looks good! Thank you!

vitcpp:
I've added one more commit to set extra_float_digits = 2 for epochprop and bounding_box_gist tests. It allows to remove expexted/bounding_box_gist_2.out but expected/epochprop_1.out is still required. The reason to create expected/epochprop_1.out is that the earlier versions output exactly 17 digits if extra_float_digits is 2, some of these digits meaningless. PG versions 12+ use a special function that outputs only meaningful digits even if extra_float-digits = 2, it is why in PG 12+ some numbers may contain less than 17 digits if the text representation if enough to define the double number.

esabol (approved):
Looks better!

PR #121, opened 2024-04-17
Alter spoint op family to add missed operator 17 (<->)

vitcpp:
The upgrade script misses the altering of spoint operator family to add operator 17. It is required for k-NN search.


df7cb:
Fwiw I think you shouldn't touch --1.5.0.sql given that has been released, even when it was just a pre-release. (I would suggest labeling pre-releases as beta or RC, then the question might not be as relevant.)

vitcpp:
@df7cb That was my unspoken question - whether to move the tag or create 1.5.1. Looking at our README I think to create 1.5.1. I will update the patch. It seems, the next release will be 1.5.1. Zeroes are unhappy :) Thank you.

esabol (approved):

esabol:
Pre-releases are not releases. I'm ok with this being part of 1.5.0, personally.

df7cb:
But git tags are tags and should not be moved if people might already have seen them. I have not yet built Debian packages from that tag yet, but if I had and you moved it around afterwards, I would be quite upset.

Also, if the pre-release is supposed to serve any purpose, it is for people to actually build and test it. Please don't confuse people by making another release with the same name.

vitcpp:
@esabol It seems, it is not easy to move the tag, the branch is protected. Furthermore, @df7cb has reasonable arguments against it. I will create 1.5.1 this time. Later, I will assign pre-releases more carefully.

esabol:
Well, Ok, but then there's no point in having pre-releases at all, and I suggest we not have them in the future.

df7cb:
Just name them 1.6rc1 or the like and everything is fine.

vitcpp:
@esabol Are you ok with these changes?

esabol (approved):
I approved the PR 3 days ago, but I just saw you made another commit. Yes, it still looks good to me.

PR #123, opened 2024-04-27
Fix build under mingw64

vitcpp:


vitcpp:
@robe2 This patch should fix your issue. Could you please check the compilation of this patch on your side? Thank you in advance.

esabol:
These changes are more extensive than I was expecting. I'm not sure they should be included in 1.5.1.

vitcpp:
@esabol Actually, my changes are in sparse.y and in sscan.l. Other files (h/c) are auto generated from those two ones.

src/sscan.c
```diff
@@ -1,6 +1,6 @@
-#line 2 "sscan.c"
+#line 2 "src/sscan.c"
```
esabol, line 1: This and the other changes which reference "src/somefile" don't look right to me. I suspect you autogenerated these files when you were in the directory above the "src" directory. Was that intentional?

esabol:
@vitcpp wrote:
> Actually, my changes are in sparse.y and in sscan.l. Other files (h/c) are auto generated from those two ones.

I realize that. It's still more changes than I'm comfortable with personally, but whatever. If you want to include it in 1.5.1, I'm fine with it.

vitcpp:
@esabol Well, lets postpone this fix until the next release. Honestly, I'm not comfortable with these changes as well. Tomorrow I will create a new release.

src/sscan.c
esabol: Ping @vitcpp

esabol: Well, if it's intentional and expected, then I'm fine with it!

vitcpp: Just some more explanation.

There are two files **sparse.y** and **sscan.l** that are used to generate (**src/sparse.h**, **src/sparse.c**) and **sscan.c** accordingly. You may remove these generated sources and run make - these files will be generated again. These files are not required to be in the repository. I think, they were added to simplify the build process - lex/yacc tools are not required.

The following commands are used by make to generate these sources:
`
/usr/bin/flex -Psphere -osrc/sscan.c src/sscan.l
yacc -d -p sphere_yy -o src/sparse.c src/sparse.y
`

Some times ago the sources were located in the pgsphere root directory. Then, we moved source files to src subdirectory. It seems, these files were not re-generated after that.  Once we run flex/yacc from the pgsphere root directory, we add src/ prefix to the input file name. As the result, #line directive is changed. It may affect error messages when the compiler finds some build errors after that directive - file and line will be taken from this directive. I do not see any other ways it may affect the functionality somehow.

Some other changes occurred when I used newer versions of lex/yacc. As the result we see some changes in comments, produced by these tools in generated files (different tool versions or urls, some new explanations). These changes should not produce any problems.

There are some new token types in sparse.h file should be investigated:
- YYEMPTY = -2
- YYEOF = 0,                     /* "end of file"  */
- YYerror = 256,                 /* error  */
- YYUNDEF = 257,            /* "invalid token"  */

I'm not sure how they affect the parser functionality. I think, I should investigate it just to make sure everything is ok (tests are passed ok with these changes).

I haven't investigated the changes in sparse.c. I think there is no sense to do it. Because lex/yacc with different versions may produce different code.

My own changes - add TOK_ prefix to some token names (SIGN, INT, FLOAT, EULERAXIS) in input files sparse.y, sscan.l to fix the compilation.

In general, we may remove the generated files and ask user to install lex/yacc to build pgsphere.

Finally, I think to investigate how these new tokens affect the parser to make sure everything is ok.

vitcpp:
I decided to propose a less invasive fix. There are two commits: first commit just regenerates parser files, second commit does the fix of token names. I used GNU Bison 3.4.2 to generate parser files.

esabol (approved):

Makefile.common.mk
```diff
@@ -4,5 +4,5 @@
 #
 #----------------------------------------------------------------------------
 
-EXTENSION        	:= pg_sphere
-PGSPHERE_VERSION 	:= 1.5.1
+EXTENSION			:= pg_sphere
```
esabol, line 7: This whitespace change doesn't look right on GitHub, but I take it looks right in your editor?

vitcpp: @esabol I just replaced spaces with tabs. It was ok for me because my editor used 4 spaces for tab. I will roll back this change. Thank you.

esabol (approved):

PR #134, opened 2024-11-02
Build and check workflow for MS Windows

vitcpp:
Add workflow for build and test for MS Windows with USE_HEALPIX = 0 only.

### DIFF
--- a/.github/workflows/build-and-check-windows-latest.yml
+++ b/.github/workflows/build-and-check-windows-latest.yml
@@ -70,6 +70,10 @@ jobs:
         pg_ctl -D pgdata -l postgres.log start
         make USE_HEALPIX=0 installcheck
 
+    - name: Show installcheck regression.diffs
+      if: ${{ failure() }}
+      run: cat regression.diffs
+
     - name: Upload artifacts
       uses: actions/upload-artifact@v4
       if: success() || failure()
--- a/.github/workflows/build-and-check-windows-latest.yml
+++ b/.github/workflows/build-and-check-windows-latest.yml
@@ -1,0 +1,81 @@
+name: Build and Check (windows-latest)
+
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * 5'
+
+jobs:
+  build_and_test:
+
+    runs-on: windows-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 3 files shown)


esabol:
Why did you remove the "Show ... regression.diffs" steps, @vitcpp ?

For what it's worth, I'd like to see a similar macOS test workflow (in a separate PR, of course).

vitcpp:
Hi @esabol, I've added the missed step. Thank you. I still failed to implement crushtest for windows. I would prefer to merge this PR without crushtest right now and come back to crushtest implementation later. This PR is not an ideal solution because it takes a lot of time to build postgresql of a particular version. I haven't found suitable packages which I can use to build pgSphere under msys2. Please, let me know if you have some ideas.

I agree about macOS, let me think about it.

esabol (approved):

vitcpp:
@esabol 

> OK. crushtest doesn't work?

Well, two tests from crushtest are failed because they run perl from psql using '\!' command to generate some input data. It seems, due to some differences in path handling between mswin and linux, it fails. I think, i will try to fight with it one day more.

esabol:
@vitcpp : Maybe try this: https://stackoverflow.com/a/60772585/5153779

vitcpp:
@esabol Thank you. I guess, it was psql issue. I did some changes in https://github.com/postgrespro/pgsphere/pull/134/commits/4922a2efe8d6e9ea5d08483265f3cc2256566fc0 and it fixed my problem. Now, I added crushtest execution.

esabol:
> @esabol Thank you. I guess, it was psql issue. I did some changes in [4922a2e](https://github.com/postgrespro/pgsphere/commit/4922a2efe8d6e9ea5d08483265f3cc2256566fc0) and it fixed my problem. Now, I added crushtest execution.

I saw that. That's why I made my suggestion. If it works, you could avoid changing sql/init_extended.sql and expected/init_extended.out.

vitcpp:
Well, I'm not sure it could help. On the screenshot below you can see the error under ms win. psql seems to interpret testsuite as an external command unfortunately. As I understood, your solution proposes to copy the perl binary to path, where it can be found with the linux shebang. I believe, the error occurs before using shebang.

image

If I change / -> \ in testsuite/gen_point..., it opens a notepad with the script :)

My solution is to add perl command which can be found in PATH variable for linux and ms win. There should be no problems on both systems.

esabol (approved):

PR #141, opened 2025-10-08
Fix Build and Check under MS Windows workflow

vitcpp:


esabol (approved):
I was just about to look into this myself. I'm glad the fix was so easy.

Issue #132, opened 2024-10-15
GitHub Actions workflows use deprecated/outdated actions


esabol:
The Actions tab shows the following warning:

The following actions use a deprecated Node.js version and will be forced to run on node20: actions/upload-artifact@v3. For more info: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

https://github.com/actions/upload-artifact says:

actions/upload-artifact@v3 is scheduled for deprecation on November 30, 2024. [Learn more.](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/) Similarly, v1/v2 are scheduled for deprecation on June 30, 2024. Please update your workflow to use v4 of the artifact actions. This deprecation will not impact any existing versions of GitHub Enterprise Server being used by customers.

So I think `uses: actions/upload-artifact@v3` should be changed to `uses: actions/upload-artifact@v4` in `.github/workflows/build-and-check.yml`.

So I thought I'd check the actions used in `.github/workflows/deploy-docs.yml` as well. Although not deprecated, `actions/upload-pages-artifact` is up to v3, and `actions/deploy-pages` is up to v4, but I'm not sure it's worth updating those?

https://github.com/actions/upload-pages-artifact?tab=readme-ov-file

https://github.com/actions/deploy-pages?tab=readme-ov-file

esabol:
OK, taken care of `actions/upload-artifact` in `.github/workflows/build-and-check.yml`.

Shall I update the actions in `.github/workflows/deploy-docs.yml` also or just leave it alone since it's working fine and there aren't any deprecation warnings for those?

vitcpp:
@esabol I can't say for sure about deploy-docs.yml. I'm not so experienced in the github actions stuff. Sometimes I try to learn it, but it seems I'm at the beginning of my long road to the understanding of the actions :) If you think to update the .yml file, I think it is ok to try. It doesn't affect the code. We can rollback it if something go in a wrong way.

esabol:
Closing this issue after the merging of PRs #133, #143, and #144.

PR #143, opened 2025-10-15
Fix Deploy Docs action

vitcpp:


esabol (approved):

esabol:
`actions/upload-pages-artifact@v4` is the most current.

Reference: https://github.com/actions/upload-pages-artifact/releases

This would close issue #132.

esabol:
Note that `actions/upload-pages-artifact@v4` has one backwards incompatible change: hidden files (specifically dotfiles) are no longer included in the artifact to be uploaded. I doubt that matters to the documentation?

esabol:
You might want to consider changing `actions/deploy-pages@v2` to `actions/deploy-pages@v4` as well.

Reference: https://github.com/actions/deploy-pages/releases/

PR #144, opened 2025-10-16
Fix doc deploy

vitcpp:


.github/workflows/deploy-docs.yml
```diff
@@ -2,9 +2,7 @@ name: Build and Deploy Docs
 
 # Deploy docs only for master
 on:
-  push:
-    branches:
-      - "master"
+  workflow_dispatch:
```
esabol, line 7: I'm all for adding `workflow_dispatch:`, but why are you removing the `push: branches: - "master"` lines?

esabol: Never mind. I get it. We want the docs to only update when there's a new release. May I suggest the following instead?
```
on:
  release:
    types: [published]
  workflow_dispatch:
```
That will trigger automatically only when a new release is published. In the context of a GitHub Actions workflow, "published" means that a release has been finalized and made visible to the public, or specifically, marked as a non-draft and non-prerelease version. This is distinct from simply saving a draft.

Reference: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#release

vitcpp:
@esabol Thank you for the idea. I've added doc deployment on release publication.

esabol (approved):

esabol:
Looks good, I hope this will get the GitHub Actions workflow to update the documentation on the website working again.

This PR will also close issue #132.

PR #145, opened 2025-10-20
Fix upgrade script generation from 1.5.1 to 1.5.2

vitcpp:
During the upgrade testing it was found, that the upgrade script was not generated. The upgrade to 1.5.2 was failed. The patch fixes the generation of the upgrade script.


esabol (approved):
Yes, I'm glad you tested an upgrade. Thank you, @vitcpp !

